PR #3, opened 2019-07-15
Completed - Please review

hrshah5:


Problem1.java
```diff
@@ -0,0 +1,14 @@
...
+        for(int i = 0; i<nums.length; i++){
+            int secondNumber = target - nums[i];
+            if(hm.containsKey(secondNumber)){
+                return new int[]{i,hm.get(secondNumber)};
+            }
+            hm.put(nums[i], i);
+        }
+        return new int[] {-1, -1};   
+    }
+}
```
ashmeetdeol19: That's pretty good Solution 👍

Problem2.java
```diff
@@ -0,0 +1,29 @@
...
+public static void main(String args[]) 
+{ 
+	int val[] = new int[]{60, 100, 120}; 
+	int wt[] = new int[]{10, 20, 30}; 
+	int W = 50; 
+	int n = val.length; 
+	System.out.println(knapSack(W, wt, val,n)); 
+	} 
+} 
+
```
ashmeetdeol19: This recursive solution seems pretty good, but you should also submit the DP solution.

PR #4, opened 2019-07-15
Please Review.

Shazmaan:


Problem1.cpp
```diff
@@ -0,0 +1,32 @@
...
+        return ret;
+    }
+};
+
+int main(){
+    Solution s;
+    vector<int> vec{2,7,11,15};
+    cout<<s.prob(vec,9).at(1)<<endl;
+    return 0;
+}
```
ashmeetdeol19: That looks like a pretty  good Solution.

Problem2.cpp
```diff
@@ -0,0 +1,39 @@
...
+        return arr[val.size()-1][weight];
+    }
+};
+
+int main(){
+    Solution s;
+    vector<int> val{60,100,120};
+    vector<int> wt{10,20,30};
+    cout<<s.prob(val,wt,50);
+}
```
ashmeetdeol19: Well Done in this one as well. 👍

PR #5, opened 2019-07-15
Completed Competitive Coding 2 - Please Review

bharatreddypothireddy:


Problem1.java
```diff
@@ -0,0 +1,24 @@
...
+// Space Complexity :O(n)  ---> because of Hashmap
+import java.util.HashMap;
+
+public class problem1 {
+    public static int[] findindex(int[] nums,int target){
+        if(nums.length == 0 || nums.length == 1)
+            return new int[]{-1,-1};
+        HashMap<Integer,Integer> map = new HashMap<>();
+        for(int i = 0;i<nums.length;i++){
+            if(map.containsKey(target-nums[i]) && i != map.get(target-nums[i]))
```
ashmeetdeol19: i != map.get(target-nums[i]) 
This condition will always be True.
Dont you think so ?

Problem2.java
```diff
@@ -0,0 +1,26 @@
...
+            }
+        }
+        return dp[weights.length][W];
+    }
+    public static void main(String[] args) {
+        int[] weights = new int[]{10, 20, 30};
+        int[] values = new int[]{60, 100, 120};
+        int W = 50;
+        System.out.println(helper(weights,values,W));
+    }
```
ashmeetdeol19: This looks like a pretty good solution.

PR #19, opened 2019-10-21
Completed the competitive coding problems please review

Abhinav-Balasubramanian:


Problem2.java
```diff
@@ -0,0 +1,46 @@
...
+        if (result.containsKey(dif) && result.get(dif) != i) {
+            return new int[] { i, result.get(dif) };
+        }
+    }
+    //If no such numbers exists an array of -1,-1 is returned
+     return new int[] {-1,-1};
+}
+}
+//Doubts with respect to this question
+// 1. While using the hashmap, the time complexity remains the same as we go through the array twice, once to load the elements in the array to hashmap and second time to find the differences.
```
sehaj-grover: Abhinav in the 2 pointer problem the complexity is O(n^2) because to check the list for every element where as using hashmap you only traverse twice no matter the input so it's 2O(n) i.e. O(n)

Problem2.java
```diff
@@ -0,0 +1,46 @@
...
+// The space complexity also increases from O[1] to O[N], so is 2 pointers the optimal solution for this problem?
+// 2. If the two numbers doesn't exists , I am returning an array of -1 but is it what we should be doing?
```
sehaj-grover: Actually it's mentioned in the problem that there will always be a pair So the last return is technically useless It's just for the compiler

Problem2.java
```diff
@@ -0,0 +1,46 @@
...
+        result.put(nums[i], i);
+    }
+    //Finding the differences of elements
+    for (int i = 0; i < nums.length; i++) {
+        int dif = target - nums[i];
+        //Checking if the differences exists in hashmap and is not the same as the element considered
+        if (result.containsKey(dif) && result.get(dif) != i) {
+            return new int[] { i, result.get(dif) };
+        }
+    }
```
sehaj-grover: can you implement using single pass instead of 2 pass

PR #49, opened 2020-03-31
Completed both problems

arcsoftech:


Problem1.java
```diff
@@ -0,0 +1,40 @@
+//Knapsack 0/1
+
+// Time Complexity :O(n)
```
devashishnyati, line 3: What is n here?

devashishnyati:
Good job Karthik!
Please verify time and space complexities for knapsack problem.
Also, submit the recursive solution for knapsack problem.

PR #53, opened 2020-04-01
Solutions for knapsack and 2sum

AroopN:


devashishnyati:
Good job Aroop!
Submit the brute force solution also.

PR #65, opened 2020-05-05
Completed Design

JacobStephen9999:


iivaishnavii:
Great job, Jacob! You got both of the solutions right.

PR #66, opened 2020-05-05
Completed problem 1

sbsreedh:


twoSum.py
```diff
@@ -0,0 +1,18 @@
+#TC:O(n^2)
```
iivaishnavii, line 1: Try out the O(n) solution as well.

iivaishnavii:
Good job,Sonali! Try out 0-1 knapsack problem too.

PR #67, opened 2020-05-05
Problem28 - Please review

KGurpreet:


Problem28 - Please review
```diff
@@ -0,0 +1,24 @@
+// Time Complexity :O(n)
```
super30admin, line 1: Good. Whats the brute force solution

super30admin:
Good. Whats the brute force solution?

Problem28 - Please review
KGurpreet: Brute force : Iterate through each element and check if there is any other value equal to( target - element)

PR #76, opened 2020-05-10
Update Problem2.java. Please review Competitive_Coding_2 Two-Sum

Abinashi7:


iivaishnavii:
Good job, Abinashi! Update TC and SC as well.

PR #133, opened 2020-07-27
Create Two Sum - Competitive_Coding_2 . Please review

firoz1905:
It is failing for input [3,2,4] . Expecting O/P - [1,2]. My O/P=[0,2] because I have sorted it. Can you pleas check

### DIFF
--- a/Two Sum
+++ b/Two Sum
@@ -1,0 +1,25 @@
+Approach 1 : Brute Force . using two loops and iterate and check the sum is equal to target
+Time Complexity : O(n*2)
+Space Complexity : O(1) -- > Result Array 
+class Solution:
+    def twoSum(self, nums: List[int], target: int) -> List[int]:
+        for i in range (len(nums)):
+            for j in range (i+1,len(nums)):
+                if (nums[j]+nums[i]==target):
+                    return i,j
+ Approach 2 : Two Pointer Approach
+ Time Complexity :O(nlogn) 
+ Space Complexity : O(1). Given result Array
+ class Solution:
+    def twoSum(self, nums: List[int], target: int) -> List[int]:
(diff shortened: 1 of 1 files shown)


Two Sum
```diff
@@ -0,0 +1,25 @@
+Approach 1 : Brute Force . using two loops and iterate and check the sum is equal to target
+Time Complexity : O(n*2)
+Space Complexity : O(1) -- > Result Array 
+class Solution:
+    def twoSum(self, nums: List[int], target: int) -> List[int]:
+        for i in range (len(nums)):
+            for j in range (i+1,len(nums)):
+                if (nums[j]+nums[i]==target):
+                    return i,j
+ Approach 2 : Two Pointer Approach
```
Sumesh-H, line 10: I guess because of sorting, it is missing the required pair. You have to take care of the indexes as well. I'm sharing my java solution refer and implement.
        int[] res = new int[2];
        int[] copy = Arrays.copyOf(nums, nums.length);
        Arrays.sort(copy);
        
        int left = 0;
        int right = nums.length - 1;
        
        while (left < right) {
            int sum = copy[left] + copy[right];

            if (sum == target) {
                break;
            } else if (sum < target) {
                left++;
            } else {
                right--;
            }
        }
        
            if (nums[i] == copy[left]) {
                res[0] = i;
            }
        }
        
        for (int i = nums.length - 1; i >= 0; i--) {
            if (nums[i] == copy[right]) {
                res[1] = i;
            }
        }
        
        return res;
    }
}

Sumesh-H:
Good Job, Firoz! Kindly check inline comment. Try the other problem as well.

Sumesh-H:
Good Job, Firoz! Kindly check inline comment. Try the other problem as well.

PR #170, opened 2020-09-30
Competitive Coding 2 PR

Jaisu-1:


ramya-sruthi-d (commented):
knapsack problem: The solution is correct. But when you are using a 2D array how will space or time complexity will be equal to O(n). Figure out them properly.
Good work, understand the approach, and explain it in our code.

ramya-sruthi-d:
knapsack problem: The solution is correct. But when you are using a 2D array how will space or time complexity will be equal to O(n). Figure out them properly.
Good work, understand the approach, and explain it in our code.

PR #193, opened 2020-10-28
competitive coding-2 submissions

PratikDhumalNetapp:
competitive coding-2 submissions

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,23 @@
+//Time O(n)
+//Space O(n)
+class Problem1 {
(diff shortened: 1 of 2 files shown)


Gupta-Keshav-001378461 (commented):
Try the single pass solution for two sum,
Knapsack problem will give you error please run your code before submission. Where is the max function you are using on line 17th.

Gupta-Keshav-001378461:
Try the single pass solution for two sum,
Knapsack problem will give you error please run your code before submission. Where is the max function you are using on line 17th.

PR #196, opened 2020-10-28
Pull request from personal id

pratikdhumal304:
Pull request from personal id

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,23 @@
+//Time O(n)
+//Space O(n)
+class Problem1 {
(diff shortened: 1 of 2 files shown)


Gupta-Keshav-001378461 (commented):
The knapsack solution is wrong on line 17 you used the max function which is not defined anywhere. Try to do the Two sum problem with single pass as well.

Gupta-Keshav-001378461:
The knapsack solution is wrong on line 17 you used the max function which is not defined anywhere. Try to do the Two sum problem with single pass as well.

PR #230, opened 2021-01-05
Competitive coding-2 done, please review

Lohita1114:


samarthjj:
Good work Lohita, Please also complete the recursive solution for the knapsack problem as well.

samarthjj:
Good work Lohita, Please also complete the recursive solution for the knapsack problem as well.

PR #273, opened 2021-03-01
both problems

ronak-prog:


khushal1198 (commented):
Missing number was the Competetive-Coding-1 problem. Please submit two sum and knapsack problem for Competetive-Coding-2

khushal1198:
Missing number was the Competetive-Coding-1 problem. Please submit two sum and knapsack problem for Competetive-Coding-2

PR #394, opened 2021-08-12
Completed competitive coding 2- Please review

sayantiroy90:


Rooppesh:
Good Job! Add the TC, SC and Approach used!

Rooppesh:
Good Job! Add the TC, SC and Approach used!

PR #637, opened 2022-07-16
Completed TwoSum and Knapsack-Please review my submission

Butulparveen:


TwoSum.java
```diff
@@ -0,0 +1,20 @@
+//Amazon Interview
+//Time Complexity = O(n)
+//Space Complexity =O(1)
```
daredevilteja, line 3: SC is O(n) because we are using a map here

daredevilteja (changes requested):
Please implement the dp solution of the knapsack problem

daredevilteja:
Please check the comments

PR #649, opened 2022-08-08
Complete `CC-2` Using Python3

ViswanathBalusu:


daredevilteja (changes requested):
Please recheck the mentioned complexities for knapsack problem

daredevilteja:
Good Work! Keep it up

ViswanathBalusu:
> Please recheck the mentioned complexities for knapsack problem

Actually TC Should be O(n^2) and SC should be O(n) where n is Max weight in the given array.
Is my approach correct?

daredevilteja:
> > Please recheck the mentioned complexities for knapsack problem
> 
> Actually TC Should be O(n^2) and SC should be O(n) where n is Max weight in the given array. Is my approach correct?

This is correct

PR #672, opened 2022-09-21
Leetcode 1 - Two Sum problem added

khushbukareliya07:


pvk3081 (commented):
Good Job. Please add a second problem also.

PR #677, opened 2022-09-27
please review

gupta-shubhangi-gs:


118. Pascal's Triangle.py
```diff
@@ -0,0 +1,27 @@
+TC:O(N)
```
daredevilteja, line 1: TC is wrong

daredevilteja:
Good Work! Keep it up

PR #687, opened 2022-10-18
Was asked the knapsack with a variation - to return the max weight instead of value

18ksangeeta:


daredevilteja:
Good Work! Keep it up

PR #806, opened 2023-10-12
 Competitive-Coding-2

sharanvodapally:


Problem1.java
```diff
@@ -0,0 +1,44 @@
+// # Competitive-Coding-2
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
ishika2741, line 3: map would require space, so space complexity here would be O(n)

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #829, opened 2024-01-03
 Competitive Coding - 2

ag-sh1996:


0-1knapsack.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(n)
```
ishika2741, line 1: Kindly check you TC and SC. Also explain what your variables actually are

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #835, opened 2024-01-05
Completed interview-2

keerthana-vijaykumar:


Problem1.cpp
```diff
@@ -0,0 +1,44 @@
+// Time Complexity : O(n log n)
```
ishika2741, line 1: Kindly submit optimized solution using hashing as well

Problem2.cpp
```diff
@@ -0,0 +1,26 @@
+// Time Complexity : O(M*N)O(N)
```
ishika2741, line 1: Kindly mention one complexity which is correct

ishika2741 (commented):
Good work!

ishika2741:
Good work!

Problem1.cpp
keerthana-vijaykumar: okay

PR #869, opened 2024-04-05
Updated Problem1 

ZeelSatasiya:


ishika2741 (commented):
Kidnly add tc and sc for this. Also submit second problem as well.

ishika2741:
Kidnly add tc and sc for this. Also submit second problem as well.

PR #906, opened 2024-07-08
completed competitive coding 2 - 2sum and knapsack 0/1

dhikki-study:
########1. Two Sum######################################################
// Time Complexity : n
// Space Complexity : n
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : though to do with 2 pointer but hasp was more efficient

// Your code here along with comments explaining your approach in three sentences only
Took hash map and than found diff of elements and stored in hashmap if found the match than return result

        

        

########https://www.geeksforgeeks.org/0-1-knapsack-problem-dp-10/#####################################
// Time Complexity : recursion - 2^number of weight, tabular - number of weightXcapacity
// Space Complexity : recursion - not idea, tabular - number of weightXcapacity
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : first did recursion than worked tabular way, tabular was having bit problem but saw previous video and got idea

// Your code here along with comments explaining your approach in three sentences only
In recursion I kept select/not select each weight and for tabular build tabular data as per class

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,88 @@
+########1. Two Sum#################################################################################################################
+// Time Complexity : n
+// Space Complexity : n
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : though to do with 2 pointer but hasp was more efficient
+
+// Your code here along with comments explaining your approach in three sentences only
+Took hash map and than found diff of elements and stored in hashmap if found the match than return result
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1021, opened 2025-02-10
CC- 2 done

gramprasadreddy:


Erlybird:
Good job! Please mention the TC and SC

Erlybird:
Good job! Please mention the TC and SC

PR #1088, opened 2025-06-11
Competitive-coding-2 complete

piyushmanglani08:


super30admin:
- The logic for checking the complement is incorrect. You should check if `diff` (the complement) exists in the hashmap, not `nums[i]`. The correct approach is to store each number in the hashmap as you iterate and check if the complement of the current number exists in the hashmap.
- The variable names are clear, and the code is easy to follow, which is good.
- Consider adding a comment to explain the approach, especially since the logic is a bit tricky.
- Here's the corrected version of your code:
  ```python
  class Solution:
      def twoSum(self, nums: List[int], target: int) -> List[int]:
          hashmap = {}
          for i in range(len(nums)):
              diff = target - nums[i]
              if diff in hashmap:
                  return [hashmap[diff], i]
              hashmap[nums[i]] = i
  ```

PR #1089, opened 2025-06-18
Solved both the problem

arihantjain2198:
Solved both the problem


super30admin:
- Problem1.java: Excellent job! Your solution is efficient and correctly implemented. The use of a hash map to reduce the time complexity is a great approach.
- Problem2.java: Your solution correctly implements the recursive approach for the knapsack problem. However, it would benefit greatly from memoization or dynamic programming to avoid recalculating the same subproblems multiple times. Consider adding these optimizations to improve the time complexity.

PR #1092, opened 2025-06-26
Merge_Sweta

swetapati22:
Merge_Sweta


super30admin:
- The student's solution for Two Sum is excellent, with optimal time and space complexity. The use of a hashmap to store complements is a smart approach.
- The knapsack solution is correct but could be improved with memoization or dynamic programming to handle larger inputs more efficiently.
- The student should ensure that the problems they are solving match the reference problem when submitting solutions for evaluation. In this case, the knapsack problem is unrelated to the Two Sum problem.
- The code quality is high, with good comments and structure. Keep up the good work!

PR #1094, opened 2025-06-30
Completed Mock Interview Questions - 2

rachanamurty:


super30admin:
Strengths:
- Good understanding of different approaches (hashmap, recursion, DP) for solving problems.
- Code is well-commented and easy to follow.
- Correctly identifies time and space complexities for each approach.

Areas for Improvement:
- In zero_one_knapsack.java, the base case logic could be simplified. The current implementation has redundant checks and calculations.
- The recursive knapsack solution could benefit from memoization to improve its time complexity.
- Some comments are overly verbose and could be more concise.
- The two_sum.java solution could handle edge cases more explicitly (e.g., empty array or no solution).

PR #1097, opened 2025-07-01
Competitive-Coding-2

mdashora:


super30admin:
Strengths:
- The solution correctly implements the problem requirements with an optimal time complexity.
- The use of a hash map for O(1) lookups is a smart approach.
- The base case check for null or empty input is a good practice.

Areas for Improvement:
- The comment "We solve this problem by iterating through array and creating map of target and index" could be more precise. It should clarify that the map stores the element and its index, not the target and index.
- The return type for the base case is `null`, which might not be ideal. It would be better to return `new int[]{-1, -1}` to maintain consistency with the reference solution and avoid potential null pointer exceptions in calling code.
- The variable `result` could be named more descriptively, such as `indices` or `pair`.

PR #1105, opened 2025-07-08

joelkny97:


super30admin:
- Exercise_1.py: Excellent job! Your solution is efficient and correctly handles the problem. The use of a hashmap is a great choice for reducing the time complexity. The code is clean and well-commented.
- Exercise_2.py: Good implementation of the knapsack problem. Your understanding of the DP approach is evident. One suggestion for improvement is to explore space optimization by using a 1D array, which could reduce the space complexity to O(W). However, your current solution is correct and efficient.

PR #1121, opened 2025-08-31
Competitive Coding 2 V1

v-hemish:


super30admin:
- The solution provided does not match the reference problem (twoSum). It's important to ensure that the solution addresses the correct problem statement.
- For the Knapsack problem, the implementation is good. The use of memoization is effective, and the code is clean and readable.
- Consider adding comments to explain the purpose of the function and the parameters for better clarity.
- If the intention was to solve the twoSum problem, a different approach would be needed, such as using a hash map for O(n) time complexity.

PR #1125, opened 2025-09-01
Competitive coding 2 Done

candida17:


super30admin:
- Problem1.java: Excellent job! Your solution is efficient and correctly implemented. The use of a HashMap to store complements is a great approach.
- Problem2.java: Your approach to the knapsack problem is on the right track with memoization. However, there's a logical error when the current item's weight exceeds the remaining capacity. Instead of returning -1, you should skip the item and proceed to the next one. Also, consider adding more comments to explain the recursive logic and memoization steps for better clarity.

PR #1127, opened 2025-09-02
knapsack, two sum

Saideepthi123:


super30admin:
- Problem1.py: The solution is excellent. The use of a hash map to reduce time complexity is a great optimization. The comments are clear and the code is easy to understand. One minor suggestion: the return type could be more consistent (e.g., returning [-1, -1] instead of -1 to match the expected return type of a list).
- Problem2.py: The solution is well-implemented and follows the standard DP approach for the knapsack problem. The comments are very helpful in understanding the logic. One suggestion: the function name `my_function` could be more descriptive, such as `knapsack` or `maxProfit`.

PR #1129, opened 2025-09-09
completed Competitive-Coding-2

aditya18m:


super30admin:
- Problem1.cpp: Excellent job! Your solution is optimal and well-explained. The use of a hashmap to reduce the time complexity from O(n^2) to O(n) is a great approach.
- Problem2.cpp: Good implementation of the knapsack problem. Your approach is correct and efficient. However, the space complexity can be further optimized to O(n) by using a 1D array instead of a 2D matrix. Also, the comment about optimizing space to O(1) is incorrect; it should be O(n). Consider revising the space optimization part of your comments.

PR #1131, opened 2025-10-13
Competitive-Coding-2 Complete

PrasiddhShah:


super30admin:
- Problem1.py: Excellent job! Your solution is efficient and correctly implemented. Just a minor correction: the space complexity is O(n) due to the hashmap, not O(1).
- Problem2.py: You have a good understanding of the knapsack problem, but there are some logical errors in your implementation. Specifically, the loop range should be `range(1, col)` instead of `range(1, bagW)`, and the return statement should reference `dp[row-1][col-1]`. Fixing these will make your solution correct. Keep up the good work!

PR #1132, opened 2025-10-13
Add knapsack and two sum solution files

ravithakur2k:
Added 01Knapsack.py implementing a memoized recursive solution for the 0/1 knapsack problem, and TwoSum.py with a hashmap-based O(n) solution for the two sum problem.


super30admin:
Strengths:
- The student's solution demonstrates a clear understanding of the problem and provides an efficient solution.
- The use of a hashmap to reduce the time complexity from O(n^2) to O(n) is excellent.

Areas for Improvement:
- While the solution is correct and efficient, it would be beneficial to handle edge cases explicitly, such as when no solution exists (though the problem likely guarantees a solution).
- The comment about space complexity could be more precise, noting that the worst-case space complexity is O(n) when all elements are stored in the hashmap.

PR #1154, opened 2026-01-10
Completed CompCoding2

JainShubham23:


super30admin:
- Strengths:
  - Problem 1: You correctly identified the optimal approach using a hashmap, which is efficient and correct.
  - Problem 2: You correctly identified the time and space complexity for the knapsack problem and attempted to implement the memoization approach.

- Areas for Improvement:
  - Problem 2: Ensure all variables (`wt`, `val`) are properly defined and passed to the methods. The `helper` method should be part of the class and properly integrated. The `return res` statement should be inside a method, not at the class level.
  - General: Always test your code to ensure it runs without errors. Pay attention to class structure and method definitions.

PR #1170, opened 2026-03-02
Adding the solutions

subbu4061:
solutions

### DIFF
--- a/Knapsack.java
+++ b/Knapsack.java
@@ -1,0 +1,19 @@
+ // TimeComplexity: O(n x W) 
+ // SpaceComplexity: O(W)
+  
+  static int knapsack(int W, int[] val, int[] wt) {
+        
+        // Initializing dp array
+        int[] dp = new int[W + 1];
+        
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The TwoSum solution is correct and efficient, with optimal time complexity.
- The code is clean and well-structured, with clear variable names and logic.
- The comments explain the approach well, which is good for readability.

Areas for Improvement:
- The submission includes an unrelated file (Knapsack.java) which is not part of the problem. In the future, ensure that only the relevant solution is submitted.
- There is a minor spelling error in the comment: "compliment" should be "complement".
- Although not necessary, you could add a check for null input or empty array, but the constraints state that the array length is at least 2, so it's acceptable as is.

Overall, the TwoSum solution is excellent. However, since the problem requires a solution to the Two Sum problem and the student included an unrelated knapsack solution, it might indicate a mix-up. But assuming the TwoSum.java is the intended solution, it is correct.

PR #1177, opened 2026-05-10
adding interview-2 solutions

nikhylw:
solution 1 (Two sum) - W2_two_sum.py
solution 2 (0/1 knapsack) - W2_01knapsack.py

### DIFF
--- a/W2_01knapsack.py
+++ b/W2_01knapsack.py
@@ -1,0 +1,30 @@
+def knapsack(W, val, wt):
+    n = len(wt)
+    dp = [[0 for _ in range(W + 1)] for _ in range(n + 1)]
+
+    for i in range(n + 1):
+        for j in range(W + 1):
+
+            # If there are no items or capacity is 0
+            if i == 0 or j == 0:
+                dp[i][j] = 0
+            else:
+                choose = 0
+
+                # Choice 1: choose the current item
+                if wt[i - 1] <= j:
+                    choose = val[i - 1] + dp[i - 1][j - wt[i - 1]]
+
+                # Choice 2: don't choose the current item
+                dont_choose = dp[i - 1][j]
+
+                dp[i][j] = max(choose, dont_choose)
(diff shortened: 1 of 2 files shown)


super30admin:
### Interview Problem : Two elements sum to target (W2_two_sum.py)
**Strengths:**
- Excellent use of hashmap to achieve O(N) time complexity, surpassing the O(N²) reference solution
- Correct algorithm logic that properly handles the complement lookup
- Good variable naming and readable code structure
- Well-commented time/space complexity analysis

**Areas for Improvement:**
1. Add `from typing import List` at the top of the file to properly import the type hint
2. Consider returning directly when found: `return [i, hashmap[complement]]` instead of appending to a result list - this is slightly more efficient and cleaner
3. The current implementation always iterates through the entire array even after finding the answer (though it returns immediately). This is a minor point.

Overall, this is a solid implementation that demonstrates good understanding of the hashmap technique for this classic problem.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem: 0-1 Knapsack Problem (W2_01knapsack.py)
**Strengths:**
- Correct implementation of the 0-1 Knapsack DP approach
- Clean, readable code with descriptive variable names
- Good use of comments explaining the logic
- Accurate complexity analysis included as comments
- Proper handling of base cases (i==0 or j==0)
- Clear separation of "choose" and "don't choose" logic

**Areas for Improvement:**
- The variable `choose` is initialized to 0, which works but could be slightly misleading. Consider initializing it conditionally or using a more explicit approach.
- The solution could benefit from a test case demonstration in the main block (like the reference solution has) to show the algorithm in action.
- While the 2D DP approach is correct, the student could explore space optimization (1D DP) as an advanced topic to reduce space complexity from O(n*W) to O(W).

Overall, this is a solid implementation that demonstrates good understanding of dynamic programming concepts.

VERDICT: PASS

PR #1185, opened 2026-06-17
solved Competitive-Coding-2

vvkumaryadla:
solved Competitive-Coding-2


super30admin:
1. **Wrong Problem**: You submitted a Knapsack Problem solution instead of the Two Sum problem. Please ensure you're solving the exact problem specified in the assignment.

2. **For Two Sum**: The classic approach uses a hash map to achieve O(n) time complexity. For each element, check if (target - current element) exists in the hash map.

3. **If you want to improve the knapsack code**: Consider optimizing space to O(n) using a 1D DP array instead of 2D, since each row only depends on the previous row.

VERDICT: NEEDS_IMPROVEMENT

---

This submission appears to be copied from a different problem entirely. The student needs to:
1. Understand the 0-1 Knapsack problem requirements: maximize value while staying within weight capacity
2. Implement a 2D DP table where dp[i][j] represents max value using first i items with capacity j
3. For each item, either skip it (dp[i-1][j]) or include it if weight allows (profit[i-1] + dp[i-1][j-weights[i-1]])
4. Use nested loops to fill the DP table
5. Return dp[n][W] as the final answer

The two-sum solution is well-written for its intended problem, but it is completely irrelevant to the knapsack problem.

VERDICT: NEEDS_IMPROVEMENT

PR #1193, opened 2026-08-12
Complete Competitive-Coding-2

tejbharath:


super30admin:
### Interview Problem: Two Sum (Problem1.java)
It appears that you have submitted a solution for the 0/1 Knapsack problem instead of the Two Sum problem. These are completely different problems. Please re-read the problem statement carefully:

- **Two Sum**: Given an array of integers `nums` and an integer `target`, return the indices of the two numbers such that they add up to `target`.
- **Knapsack**: Given weights and values, find the maximum value that can be carried in a knapsack of capacity W.

For Two Sum, you need to:
1. Iterate through the array
2. For each element, check if there's another element that complements it to reach the target
3. Return the indices of those two elements

A simple approach would be to use nested loops (O(n²) time, O(1) space), or for better efficiency, use a HashMap to store seen values and their indices (O(n) time, O(n) space).

Please resubmit with a correct solution to the Two Sum problem.

VERDICT: NEEDS_IMPROVEMENT

---

Return nonul

VERDICT: NEEDS_IMPROVEMENT

